PR #1, opened 2019-07-08
completed please review

PremalSamale:


Problem1.java
```diff
@@ -1 +1,40 @@
...
+class Problem1 
+{ 
+static int search(int ar[], 
+				int size) 
+{ 
+	int a = 0, b = size - 1; 
+	int mid = 0; 
+	while ((b - a) > 1) 
+	{ 
+		mid = (a + b) / 2;
```
super30admin: Instead of waiting for b-a to be =1, Can you thing of some early condition to terminate the loop?

PremalSamale: if(mid == a) break;

ashmeetdeol19: This Solution wont work for the given test case.
Input : [1,3,4,5,6,7,8]

PR #2, opened 2019-07-08
Please review my problem 1

AkhilKrishnamurthy:


Problem1.java
```diff
@@ -1 +1,46 @@
...
+	
+	int binarySearch(int arr[]) 
+    { 
+        int l = 0, r = arr.length - 1; 
+        int m = l + (r - l) / 2; 
+        System.out.println(m);
+        int left = m-1, right = m+1;
+        if(arr[left]!=arr[m]-1) return arr[m]-1;
+        if(arr[right]!=arr[m]+1) return arr[m]+1;
+        while (l <= r) {
```
super30admin: This condition is always true as you are never updating l and r.
We have already discussed the solution.Work on that.
Good Luck.

Problem2.java
```diff
@@ -1 +1,70 @@
...
+	class Node {
+		int data;
+		Node next;
+		Node(int d) {
+			data = d;
+			next = null;
+		}
+	}
+	
+	public void push(int new_data)
```
super30admin: You are not at all using priority in your code.
This wont work as priority queue.
Try again and submit PR once again.

PR #4, opened 2019-07-08
Completed Problem Set: Kindly review

Ameya145:


Problem1.java
```diff
@@ -1 +1,39 @@
...
+class Solution
+{
+
+	//returns missing element(if found), else -1
+	static int findMissing(int nums[])
+	{
+		int i=0;
+		for(i=0;i<nums.length;i++)
+		{
+			if(nums[i]!=i+1) //check if the element of the array is equal to the value that should be present at that index
```
super30admin: What if array starts from number other than 1. Can you think of solution to that as well?

Problem2.java
```diff
@@ -1 +1,96 @@
...
+		obj.push(2,5);
+		System.out.println(obj.peek());
+		obj.push(3,3);
+		System.out.println(obj.peek());
+		obj.pop();
+		System.out.println(obj.peek());
+		obj.pop();
+		System.out.println(obj.peek());
+	}
+}
```
super30admin: It fails for this test case. Kindly look to it.

		obj.push(4,4);
		System.out.println(obj.peek());
		obj.push(1,1);
		System.out.println(obj.peek());
		obj.push(3,3);
		System.out.println(obj.peek());
		obj.pop();
		System.out.println(obj.peek());
		obj.pop();
		System.out.println(obj.peek());

Problem1.java
```diff
@@ -32,7 +57,8 @@ static int findMissing(int nums[])
 	
 	public static void main(String args[])
 	{
-		int nums[] = {1,2,3,4,5,6,7,8,9,10};
+		int nums[] = {1,2,3,4,5,6,7,9};
+		System.out.println(findMissingBinary(nums));
 		System.out.println(findMissing(nums));
 	}
```
super30admin: It works totally fine.

PR #20, opened 2019-08-25
Create Big_N_10_problem

neha-0507:
@kaurramanpreet

### DIFF
--- a/Big_N_10_problem
+++ b/Big_N_10_problem
@@ -1,0 +1,15 @@
+class Solution(object):
+    def missingNumber(self, nums):
+        """
+        :type nums: List[int]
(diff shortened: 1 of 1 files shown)


Big_N_10_problem
```diff
@@ -0,0 +1,15 @@
+class Solution(object):
+    def missingNumber(self, nums):
+        """
+        :type nums: List[int]
+        :rtype: int
+        """
+        nums.sort()
```
kaurramanpreet: is there a need to sort the elements?

neha-0507:
Yes in leet code the list is not sorted so i sorted it first and then done
the code

PR #24, opened 2019-08-27
Completed Finding Missing element and Priority Queue - Please Review

PoojaR24:


Problem1.java
```diff
@@ -1 +1,48 @@
...
+            {
+                // the current mid as the new left index
+                l = mid;
+            }
+            //if the difference between the left and the right indices is equal to 1
+          	if(r-l ==1)
+            {
+              //It means we have reached an array of length 2 and the missing value is in between the 2 numbers hence return the average of the 2 numbers as the missing value.
+              //System.out.println(r);
+              return ((ar[r]+ar[l])/2) ;
```
kaurramanpreet: great approach!!

Problem2.java
```diff
@@ -1 +1,137 @@
...
+    //left
+    private int left(int pos)
+    {
+        return (2*pos);
+    }
+
+    //right
+    private int right(int pos)
+    {
+        return (2*pos)+1;
```
kaurramanpreet: also give explanation that why is this giving my right child 
so that i get to know that you are also aware of the logic behind it

PR #25, opened 2019-10-13
Prob-1 solution added

akhileshborgaonkar:


Problem1.java
```diff
@@ -1 +1,25 @@
...
 
+class Solution {
+    public int missingNumber(int[] arr) {
+        int low = 0, mid = 0;
+        int high = arr.length;
+
+        //Arrays.sort(arr);		//if arrays are not sorted then
+        while(low < high){
+            mid = low + (high-low)/2;
+            if(arr[mid] > mid)		//checking if the mid index is in sync with the mid index if not then probably an element in the left is missing
```
sehaj-grover: akhilesh  arr[mid] will always be greater than mid because the index starts from 0 and values start from 1

PR #38, opened 2019-12-30
Competitive Coding Problem 1

SiddharthMehtaRajendra:
Competitive Coding Problem 1. Min Heap will do after discussion in class.


rajatkumar01:
Nice try with brute force solution. As this uses the internal java function. Try to solve it with BS. As we have to search for the element and as numbers are in sorted order what we will be doing is we will be comparing the value at a particular index with its index if they dont satisfy the relation that value=index+1 then that means there is something wrong on its left so in this way we will proceed .

PR #51, opened 2020-02-11
Finished Missing Element and MinHeap, please review

sowmyadvn:
Hi,

Finished both the questions from mock interviews 1st session. Please review and let me know if something can be improved. Also, min heap space complexity is O(1) auxilliary right? 

Thanks,
Sowmya


nikrasiya:
Great job. Yes space is right.

PR #54, opened 2020-02-26
done with both, please review

singhperryo1:


singhperryo1:
MockInterview: Time Complextiy: O(logN) where N is # of elements in array and Space Complexity: O(1) 
MinHeap: Time Comlexity: O(logN) where N is number of elements in heap array. 

Thank you!

PR #56, opened 2020-03-23
Please review

AMAbhi:
Missing Number Problem


devashishnyati:
Nice try Abhishek!
You can also iterate through the array and check differences for an O(n) solution.
Try to optimize it on time by using binary search.

PR #64, opened 2020-04-27
Completed finding missing element problem in O(n)

Akshay-Ravi:


super30admin:
This is fine but can we do it in log(n) time?

Akshay-Ravi:
We can do it in O(logn) using binary search definitely, but that is if the array is sorted right? The question does not mention that the array will be sorted.

PR #70, opened 2020-04-28
Completed Day-6 PR-Please Review

sbsreedh:


missingNum.py
```diff
@@ -0,0 +1,9 @@
+#Time Comlexity:   O(n)
```
iivaishnavii, line 1: Nice!There is a binary search solution as well.. Do take a look.

iivaishnavii:
Great job, Sonali!

PR #72, opened 2020-04-28
Completed Competitive Coding - 1: Please Review

anikpait:


MinHeap.java
```diff
@@ -0,0 +1,99 @@
...
+            if (pos >= (currentSize / 2) && pos <= currentSize) {       // If leaf node is reached we are done comparing
+                return true; 
+            } 
+            return false; 
+        } 
+        public int getMin(){
+            return Heap[0];                 // Root will carry the least element always
+        }
+        
+        public void heapify(int pos){
```
super30admin, line 53: Good

MinHeap.java
```diff
@@ -0,0 +1,99 @@
...
+                if(Heap[left] > Heap[right]){
+                int t = Heap[pos];
+                Heap[pos] = Heap[right];                // Compare the parent with children and replace the smallest child with parent
+                Heap[right] = t;
+                } else {
+                int t = Heap[pos];
+                Heap[pos] = Heap[left];
+                Heap[left] = t;
+                }
+                heapify(pos);                       // Continue heapify till the heap values are completely adjusted
```
super30admin, line 69: Correct

super30admin: What is the time complexity of heapify function alone?

Missing.java
```diff
@@ -0,0 +1,28 @@
+/* Time Complexity : O(logn)-> Binary search
+// Space Complexity : O(1) iterative
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : I was doing Binary search on both sides once we get the mid, and then was checking whether a
```
super30admin, line 4: good

Missing.java
```diff
@@ -0,0 +1,28 @@
...
+//                                           arr[mid] - mid >0 but then I realised what if the start value is 1. And Binary Search on both sides is = O(n)
+/*Algorithm: Find the mid element. Compare the a[mid]- index(mid) with the start value and end value to see which side of the array is sorted.
+It means the missing number is on the unsorted side. Hence, move the start and end pointer accordingly. Continue the same till the difference
+between the element and index becomes unequal plus the end dominates the start by 1, return the avg between the two elements. 
+*/
+class Solution {
+    public int missingNumber(int[] nums) {
+        Arrays.sort(nums);
```
super30admin, line 12: This right here is O(nLogn). We are given a sorted array only

super30admin:
Good Commitment. Please review inline comments

MinHeap.java
anikpait: The heapify function gives O(n) time complexity as per my read. I have a question here that I am trying to understand how the heap building takes O(n). Now, consider I am deleting an element. I am calling heapify once I delete and element, and then heapify takes over to adjust the heap in O(log n). So total deletions are n elements and for each element, I call heapify 'n' times and it adjusts the heap in O(log n) => complexity O(nlogn) (similar to heap sort).  But I am not able to understand why do we say heapify takes O(n) since we are comparing the top element with children and moving down to leaf which takes log n steps right? I have been referring to this article: https://stackoverflow.com/questions/9755721/how-can-building-a-heap-be-on-time-complexity but not able to understand the logic behind O(n) behind heapify (ignore the maths part as of now). Can you please give me a brief explanation to my question?

Missing.java
anikpait: Oh yes, my bad! I did it in leetcode, in the question, they gave an unsorted array.

super30admin:
geeksforgeeks.org/time-complexity-of-building-a-heap/ review this

PR #75, opened 2020-04-29
Completed Missing Element Problem - Please Review

karthikmk:


sowmyadvn:
Solution for missing element problem is correct. Please try the other CC problem as well

sowmyadvn:
Solution for missing element problem is correct. Please try the other CC problem as well

PR #78, opened 2020-04-29
Update Problem1.java

shalinych:


iivaishnavii:
Good job, Shalini. Complete the other problem as well.

PR #92, opened 2020-06-01
Problems. Please Review

soumitra9:


missing_element.py
```diff
@@ -0,0 +1,22 @@
...
+# Time Complexity : O(n)
+# Space Complexity : O(1)
+# Did this code successfully run on Leetcode : Yes
+# Any problem you faced while coding this : Yes
+'''
+for given numbers in the range 0 to n, we know in advance the summation of the n natural numbers is n(n+1)/2,
+so here, subtarct the sum of given elements and and subtract it from the actual sum, and that is the missing value.
+'''
+
+class Solution:
```
Lohith-Aswathappa, line 11: Try the binary search for this problem

Lohith-Aswathappa:
Well done, Soumitra!

try to optimize the missing number problem.

missing_element.py
soumitra9: But it was not given in the question that the array was sorted. Otherwise, I would have definitely used BS first. Anyway, I will update the code.

Thanks

PR #95, opened 2020-06-01
solved missing element

Akash-Akki:
Solved missing element


Problem1.java
```diff
@@ -1 +1,33 @@
 
+// Time Complexity : O(log(n)
```
chiragshahi, line 2: define n

chiragshahi:
Also try the other question

PR #97, opened 2020-06-01
Problem 1 Solution

akshatakulkarni98:
Kindly Review


AbhiniveshP:
The solution provided is a brute force approach. Please try doing it using Binary Search. Also, kindly complete the other question as well.

AbhiniveshP:
The solution provided is a brute force approach. Please try doing it using Binary Search. Also, kindly complete the other question as well.

PR #98, opened 2020-06-01
Cometitve-Coding-1 Complete

Keerti995:


Problem2.java
```diff
@@ -1 +1,39 @@
+/**
```
chiragshahi, line 1: Try this question if the array is unsorted as well

chiragshahi:
Excellent work on both the problems! Good to see bit operations and comments in the code.

PR #101, opened 2020-06-01
Completed Competitive-Coding-1 - Problem 1 

GreeeshmaReddy:


chiragshahi:
Also try the other question

PR #106, opened 2020-06-01
Competitive coding commit

PadmaPriyaCheruvu:
Problem1 - Find missing element
Problem2 - MinHeap


chiragshahi:
Good job on both the problems! Try this question if the array is unsorted as well: https://leetcode.com/problems/missing-number/

PR #109, opened 2020-06-02
Completed : Please Review

ashiranka:


Problem2.java
```diff
@@ -1 +1,116 @@
+//Min Heap
+//Time Complexity : O(logn) for insertion and removal; O(1) for getMin
+//Space Complexity : O(1), Maximum size set to 15
```
chiragshahi, line 3: this is an input parameter on which the space complexity depends. So it is O(n) where n is the maximumsize

chiragshahi:
Great work on both the problems! Try question1 if the array is unsorted. Leetcode - Missing Number

PR #130, opened 2020-07-14
Completed: Problem 1

akashkshekarappa:
Find Missing number in sorted array using Binary Search


samarthjj:
Good work Akash, Please complete the other problem **Design MinHeap** as well.

PR #170, opened 2020-08-27
Problem-1 Solution

ManavAhlawat:


Himanshu-Agarwal93:
Correct implementation

PR #184, opened 2020-09-21
Missing number solution (Python)

kovurivinay:


Lohith-Aswathappa (commented):
Good work, Vinay!
Try the design problem as well.

Lohith-Aswathappa:
Good work, Vinay!
Try the design problem as well.

PR #200, opened 2020-09-23
Problem1 completed

yogesh696ksingh:


Lohith-Aswathappa (commented):
Good work, Yogesh!
Try design problem as well.

Lohith-Aswathappa:
Good work, Yogesh!
Try design problem as well.

PR #205, opened 2020-10-19
BS problem review in cc

maithrreye:


Solution.py
```diff
@@ -0,0 +1,29 @@
+"""
+Time Complexity Log n 
+Space complexity O(1)
```
kannan-mrugapphan, line 3: Shouldn't the time be nlogn as you are sorting

kannan-mrugapphan (commented):
Good Job. Please submit the other problem as well.

kannan-mrugapphan:
Good Job. Please submit the other problem as well.

Solution.py
maithrreye: Yeah write I forgot the sorting part . Thanks for pointing it out

PR #226, opened 2020-11-24
Problem1.java - find missing number

rishiraj824:


khushal1198 (commented):
Good job Rishi!

khushal1198:
Good job Rishi! Try to solve for the problem where the numbers are not starting from 1 and generalize the solution.

PR #242, opened 2020-12-29
Completed Probllem1

anshul-s30temp:


maithrreye:
Good Job on first problem. Please try the other problem as well. Kindly add TC and SC.

maithrreye:
Good Job on first problem. Please try the other problem as well. Kindly add TC and SC.

PR #243, opened 2020-12-29
Solved Missing Number Question

deepashree5594:


Himanshu-Agarwal93 (commented):
In the function you have passed `arr` but while returning you are using  `nums`. Approach is correct

Himanshu-Agarwal93:
In the function you have passed `arr` but while returning you are using  `nums`. Approach is correct

deepashree5594:
My mistake. I will correct that.
Thanks

PR #247, opened 2020-12-29
Update Problem1.java

shamali21:
Provide an array of n-1 integers in the range of 1 to n with no duplicates in list, One of the integers is missing in the list. Write a code to find the missing integer.


ramya-sruthi-d (commented):
Good work, mention sc as well.

ramya-sruthi-d:
Good work, mention sc as well.

PR #252, opened 2020-12-30

Pooja0730:
Please review


ramya-sruthi-d (commented):
Good job, the tc for find missing num would be o(long) since you are doing binary search.

ramya-sruthi-d:
Good job, the tc for find missing num would be o(long) since you are doing binary search.

ramya-sruthi-d (commented):
Code is fine, the tc for findmissingnum problem is O(logn) because you are using binary search. Mention the complexities of the other problem as well.

ramya-sruthi-d:
Code is fine, the tc for findmissingnum problem is O(logn) because you are using binary search. Mention the complexities of the other problem as well.

PR #277, opened 2021-01-25
Completed Sprint Problem1 - Please review

JanhaviDahihande:


Problem1.py
```diff
@@ -0,0 +1,21 @@
...
+Find missing element in sorted array in range 1 to n
+Time complexity: O(logn)
+Space complexity: O(1)
+'''
+class Solution:
+    def findMissingElement(self, arr):
+        low = 0
+        high = len(arr) - 1
+        while low < high:
+            mid = low + (high - low)/2
```
sibalnirbhay, line 11: Great work on getting the TC and SC right. Please mention the approach in your comments as well.

Also, `mid = low + (high - low)/2` would give you a floating point number.

Problem2.py
```diff
@@ -0,0 +1,40 @@
...
+    def getMin(self):
+        # Time complexity: O(1)
+        return self.myHeap[0]
+        
+    def extractMin(self):
+        # Time complexity: O(logn)
+        return heappop(self.myHeap)
+        
+    def printHeap(self):
+        print(self.myHeap)
```
sibalnirbhay, line 25: You are not supposed to use the `heapq` module to implement MinHeap. Please implement MinHeap using Arrays only.

sibalnirbhay:
`Problem1.py`:

> Great work on getting the TC and SC right. Please mention the approach in your comments as well.
> 
> Also, `mid = low + (high - low)/2` would give you a floating point number.

`Problem2.py`:

> You are not supposed to use the `heapq` module to implement MinHeap. Please implement MinHeap using Arrays only.

PR #284, opened 2021-02-17
cc 1

HarshaVardhan-Kaki:


min_heap.py
```diff
@@ -0,0 +1,18 @@
...
+        return heapq.heapify(array)
+    
+    def add(self,element):
+        heapq.heappush(self.minheap,element)
+    
+    def remove(self):
+        return heapq.heappop(self.minheap)
+
+
+
```
sibalnirbhay, line 18: You are not supposed to use `heapq` for this exercise. Please resubmit the solution to this problem.

sibalnirbhay (commented):
Read inline comments.

sibalnirbhay:
Read inline comments.

PR #288, opened 2021-02-22
Update Problem2.java

SourabhSanghi:


Problem2.java
```diff
@@ -1 +1,35 @@
 
+class GFG {
+	public static void main (String[] args) {
+		System.out.println("GfG!");
```
yahska-ednihs, line 4: Good Job! Saurabh. Please add the second problem as well.

PR #319, opened 2021-04-26
Solved Mock interview question 1: Please Review

SaurabhEdge:
During the mock interview, my one edge case was not working. I have added the same.


ramya-sruthi-d (commented):
Good job, explain your approach as well in the comments.

ramya-sruthi-d:
Good job, explain your approach as well in the comments.

PR #383, opened 2021-06-27
Completed the Assignment please review

tiwariarjun91:
I had completed this one month back but forgot to create a pull request. I apologize in advance!!!


Rooppesh:
Good Job! Please add the TC and SC! Also try the other problem as well!

Rooppesh:
Good Job! Please add the TC and SC! Also try the other problem as well!

PR #385, opened 2021-06-29
Pr for Mock Interview 1

sairamya27:


super30admin (commented):
TC and SC missing. Attempt Min Heap as well

PR #394, opened 2021-06-30
Competitive Coding - Please Review

kpathi:


AmruthashreeAM (commented):
The solution seems to be not working. Please re-check

AmruthashreeAM:
The solution seems to be not working. Please re-check

AmruthashreeAM (commented):
Good work! You may try this solution without using queue.

AmruthashreeAM:
Good work! You may try this solution without using queue.

kpathi:
Hi, @AmruthashreeAM Thank you for the comments. I am able to execute both Min Heap and Missing Element problems on Geeks for Geeks IDE and here is the result below. Thank you.

**1.) Missing Element:**
Input:  int[] arr = {1, 2, 3, 4, 5, 6};

Output:
missingElement : 7

**2.) Min Heap:** 
Head value using peek function:10
The queue elements:
10
30
20
400
Polling an element with poll function: 10
After removing an element with poll function:
20
30
400
after inserting 60 with insert function:
20
30
400
60
Polling an element with poll function: 20
After removing an element with poll function:
30
60
400

AmruthashreeAM:
Try your solution in leetcode, so that you will have all types of input covered. 
https://leetcode.com/problems/missing-number/
I m thinking this is the problem given, please let me know if i am wrong

and for minheap, your solution is correct. my suggestion was to try without using queue

kpathi:
@AmruthashreeAM The problem on leetcode is different as its input is from [0...n] and the one we were given is in the range of [1...n], please find the source link below.
Source Link: https://www.geeksforgeeks.org/find-the-missing-number-in-a-sorted-array/  

I shall try implementing minheap without using queue too. Thank you.

AmruthashreeAM:
okay I see, also the leetcode question has unsorted array as input. Sorry about the consfusion.

yeah you can implement minheap that way, thanks.

kpathi:
No problem, thank you.

PR #454, opened 2021-09-07
completed sprint Competitive coding 1_ Please review problem1

aashishkumar96:


Stnagam:
Please add TC SC and do the other one.

Stnagam:
Please add TC SC and do the other one.

PR #464, opened 2021-09-30
Completed Missing Number- Please Review

vishwaparikh:
Que : Didnt understand how it will return 1 even if 8 is missing as the last element? 
Will it always return one if there is no missing number in between? 
I didnt get that part in interview also.
Can you @jaspindersingh83  explain here? How to explain that in interview.


pkariya1 (commented):
Good Job Vishwa, it is given that that number to searched in not 1 or last Number

pkariya1:
Good Job Vishwa, it is given that that number to searched in not 1 or last Number

vishwaparikh:
@pkariya1  that was not the que I was asked in interview. She asked me how to find if end or start element is missing that's where I got stuck.

PR #471, opened 2021-10-13
Create missingnum.py

rucha2997:


Stnagam:
Do the remaining and add TC SC

PR #503, opened 2021-12-27
chore: provided optimal solution for both the problems

BitLegion1996:


Stnagam:
Please add comments, TC SC to all the problems!

Stnagam:
Please add comments, TC SC to all the problems!

PR #504, opened 2021-12-27
Completed probs #15 and 16

LiberoPat:


brunda4 (commented):
Provide complexities for missing integer problem.

brunda4 (commented):
Good Work!

brunda4:
Provide complexities for missing integer problem.

PR #534, opened 2022-01-14
Completed Competitive Coding - Please review

alexanderjac:


Missing Number in a sorted Array.py
```diff
@@ -0,0 +1,16 @@
+# Time Complexity : O(logN)
+#  Space Complexity :O(1)
+def Search(arr, size):
```
manojvarmarudraraju, line 3: This is a good solution, also try to make use of index to find the missing number. See the relation between index and missing number!!

manojvarmarudraraju (commented):
Good job, try the other problem as well!!

manojvarmarudraraju:
Good job, try the other problem as well!!

PR #575, opened 2022-03-12
Competitive coding-1 completed - Please review

suyog30:


FindMissingNumber.py
```diff
@@ -0,0 +1,21 @@
+# Time Complexity : O(logn)
+# Space Complexity : O(1)
+# Did this code successfully run on Leetcode : Yes
+# Any problem you faced while coding this : No
+
+# Perform binary search. If value at index -minus value at mid is not consistent then that means some number has gone missing.
+
```
siddhimanoj, line 7: instead just check the value at mid if it is less than target move to right side else left side

MinHeap.py
```diff
@@ -0,0 +1,33 @@
+# Time Complexity : O(logn) for insert, O(logn) for extracting min and O(1) for getMin
```
siddhimanoj, line 1: Try to build the heap from scratch!!

siddhimanoj (commented):
Good job, please build the heap from scratch!!

siddhimanoj:
Good job, please build the heap from scratch!!

PR #636, opened 2022-05-31
Please review

YashTike:
I was not able to do the second program for CC-1.

### DIFF
--- a/Problem1.cpp
+++ b/Problem1.cpp
@@ -1,1 +1,66 @@
+/*
+This program finds the missing number using binary search.
+
+Problems with the code: Code executed, but did not return answer
+*/
+
+
+#include<bits/stdc++.h>
+#include<vector>
+
+using namespace std;
+
+class Problem1
+{
+public:
+    
+    int elementFind(vector<int> vect)
+    {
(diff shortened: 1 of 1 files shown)


Problem1.cpp
```diff
@@ -1 +1,66 @@
...
+            {   
+                mid = low + (high - low)/2;
+                //Checking middle element for mismatch. If no mismatch then we search upper half
+                if(vect[mid] == mid+1)
+                    low = mid+1;
+
+                else
+                {   //Checking if element at the middle is the mismatch. If not then we search lower half
+                    if(vect[mid-1] == mid)
+                        return vect[mid]-1;
```
jeelpatel18, line 38: I checked the returning value by printing it here, except your test case it's working

jeelpatel18 (commented):
Good work. Please see the recording of solution which is shared for MinHeap problem

jeelpatel18:
Good work. Please see the recording of solution which is shared for MinHeap problem

Problem1.cpp
YashTike: Thanks! I will check it out

PR #665, opened 2022-06-28
Mock Interview One

SreekarManchukonda:


niharikagupta2021:
Good Job! Please add sc and tc as well.Complete the other problem as well

PR #685, opened 2022-07-15
Completed Competitive-Coding-1-Please Review

sahithi6445:


Problem1.java
```diff
@@ -1 +1,77 @@
+// Time Complexity :put->O(logn),delete->O(logn),where n is no of elements
+// Space Complexity :constant as we are not using any extra space
```
daredevilteja, line 2: For this problem the SC might be O(1) because you are using a constant to set the int[] space but in general the SC is O(n)

daredevilteja (commented):
Good Work! Keep it up

daredevilteja:
Good Work! Keep it up

PR #729, opened 2022-09-14
Completed both problems for CC-1

ysoni0303:


pvk3081 (commented):
Good job yash

pvk3081:
Good job yash

PR #738, opened 2022-09-21
Two sum solution

kundankarthik77:


daredevilteja:
Please complete all the problems and then raise the PR

PR #750, opened 2022-10-13
Solved the Question I knew about

18ksangeeta:
Could you please post the Problem 2 question


daredevilteja:
Implement a min heap is the 2nd problem

PR #852, opened 2023-07-15
Missing number in a sorted array problem solved

Erlybird:


poojaGit95 (commented):
Good job. Briefly explain the approach used, also try the binary search approach

poojaGit95:
Good job. Briefly explain the approach used, also try the binary search approach

PR #873, opened 2023-09-05
FindMissingElement - Please review

jailashah:
What is the 2nd problem statement?


Nandha2Tech (commented):
Great Work!
In my opinion, if array start with 3 and so on we won't get the desired output. Try to remove the first two if conditions so that it will work for input of array that doesn't starts with 1.

Nandha2Tech:
Great Work!
In my opinion, if array start with 3 and so on we won't get the desired output. Try to remove the first two if conditions so that it will work for input of array that doesn't starts with 1.

PR #885, opened 2023-10-01
Competitve coding 1 completed : Please review

ApekshaKhilari:
Question:

Find missing number in a given sorted array where range of the array is 1 to n 
At a time there will be only one missing element.


pvk3081 (commented):
Good Job Apeksha

pvk3081:
Good Job Apeksha

PR #886, opened 2023-10-01
Added changes to the code to solve the problem

nithind-sudo:


ishika2741 (commented):
Good work! Kindly add TC and SC as well.

ishika2741:
Good work! Kindly add TC and SC as well.

PR #915, opened 2023-11-15
Completed Problem 12

BhagAneesh:


ishika2741 (commented):
Good work! Kindly add Tc and Sc as well and also submit other problem as well.

ishika2741:
Good work! Kindly add Tc and Sc as well and also submit other problem as well.

PR #924, opened 2023-12-24
Completed CP-1

galibhaskar:
Find Missing Number in Sorted Array


Nandha2Tech (commented):
Great Work! Kindly include TC and SC.

Nandha2Tech:
Great Work! Kindly include TC and SC.

PR #929, opened 2023-12-26
competitive coding 12 done

Dibyajyoti227:


ishika2741 (commented):
Good work! Kindly add TC and SC as well. Also submit the second problem.

ishika2741:
Good work! Kindly add TC and SC as well. Also submit the second problem.

PR #931, opened 2023-12-26
Competitive-Coding-1 completed

dalvirushikesh:
tried solution in video in java file and then written same logic in python


pvk3081 (commented):
Good Job Rushikesh

pvk3081:
Good Job Rushikesh

PR #934, opened 2023-12-27
completed Competitive-Coding-1

ag-sh1996:


Problem1.java
```diff
@@ -1 +1,34 @@
 
+/*
+ * Time complexity : O(log n)
+ * space complexity : O(n)
```
ishika2741: Check space complexity

ishika2741 (commented):
Good work! Kindly submit solution for second problem as well.

ishika2741:
Good work! Kindly submit solution for second problem as well.

Problem1.java
ag-sh1996: Updated to O(1)

PR #957, opened 2024-01-30
Missing element and Min Heap

kpranavreddy01:


Nandha2Tech (commented):
In Find Missing Element, you are doing a linear search in line 13. Instead try to apply binary search or else TC won't be log n.

Nandha2Tech:
In Find Missing Element, you are doing a linear search in line 13. Instead try to apply binary search or else TC won't be log n.

PR #1044, opened 2024-09-25
implement problem 1

keyur-patel-21:


Erlybird:
The logic is incorrect. The array is not sorted in the question. Please correct your code and resubmit

Erlybird:
The logic is incorrect. The array is not sorted in the question. Please correct your code and resubmit

PR #1087, opened 2024-12-31
Adding java files

rahesh808:


Erlybird:
First problem is wrong, please cross check and resubmit

Erlybird:
First problem is wrong, please cross check and resubmit

PR #1122, opened 2025-01-26
submit CC-1 problem:

shreyarajwade13:


Erlybird:
Solution is incorrect, you are not handling all the edge cases, the base case is also incorrect, if the nums length is 0, you must return 1 , not -1.
There's also a logical error in the binary search. Please cross-check and resubmit

Erlybird:
Solution is incorrect, you are not handling all the edge cases, the base case is also incorrect, if the nums length is 0, you must return 1 , not -1.
There's also a logical error in the binary search. Please cross-check and resubmit

PR #1126, opened 2025-01-27
Mock interview Solution Done

gramprasadreddy:


Erlybird:
Solution is incorrect, there's a possibility of failing if the missing number is adjacent to mid, which you are overlooking by updating high = mid + 1.

Erlybird:
Solution is incorrect, there's a possibility of failing if the missing number is adjacent to mid, which you are overlooking by updating high = mid + 1.

PR #1179, opened 2025-03-20
completed problem 1 only

Mrunal401:


AdityaGudimelaS30TA (commented):
good work with all code changges.

AdityaGudimelaS30TA:
good work with all code changges.

PR #1196, opened 2025-05-19
Completed Competitive Code Problems

priyayeni:
for problem 2:

i did not use online editor to check if it is working or not, i have used notepad to write the solution.


Nandha2Tech (commented):
MissingNumber problem is fine. But in case of Design Min Heap, the logic of building the heap seems fine with some syntax errors and few bugs. Parent index can be accessed with (n-1)/2 not n/2 and initializing n=0 always won't work.

Nandha2Tech:
MissingNumber problem is fine. But in case of Design Min Heap, the logic of building the heap seems fine with some syntax errors and few bugs. Parent index can be accessed with (n-1)/2 not n/2 and initializing n=0 always won't work.

PR #1245, opened 2025-06-23
feat: find missing number in sorted array

anuragparla:
This PR contains the solution to find missing number in a sorted array.


super30admin:
Strengths:
- The code is well-structured and follows a class-based approach, which is good for modularity.
- The initial checks for edge cases (first and last elements) are correctly implemented.
- The use of binary search is appropriate for this problem, and the mid calculation is correctly done.

Areas for Improvement:
- The logic inside the while loop is incorrect. The conditions should be mutually exclusive (using `elif` instead of separate `if` statements) to avoid overlapping checks.
- The return statement should be outside the while loop to ensure the loop completes its execution.
- The loop condition should be `while (high - low) > 1` to ensure the loop terminates when the missing number is found.
- The final return statement should return `arr[low] + 1` after the loop completes.

Example corrected logic:
```python
while (high - low) > 1:
    mid = low + (high - low) // 2
    if (arr[low] - low) != (arr[mid] - mid):
        high = mid
    else:
        low = mid
return arr[low] + 1
```

PR #1252, opened 2025-06-25
Completed competitive problems

sethu95:


super30admin:
- Strengths:
  - The binary search implementation in Problem1.java is clean and handles edge cases well.
  - The min-heap implementation in Problem2.java is thorough, with clear methods for insertion, removal, and heap maintenance.
  - Good use of comments to explain the approach and logic.
- Areas for Improvement:
  - In Problem1.java, the initial check `if (nums[0] != 1) return 1;` could be generalized to handle cases where the array starts with a number other than 1, though the current implementation works for the given test cases.
  - In Problem2.java, the `size` variable is managed both as an instance variable and passed around in the main method, which could lead to confusion. It would be better to encapsulate all heap operations within the class and manage size internally.
  - The `bubbleDown` method in Problem2.java could be optimized to first find the smallest child before swapping, rather than checking left and right children separately.

PR #1275, opened 2025-07-31
Completed Competitive-Coding1

AneerudhS27:


super30admin:
Strengths:
- Clear implementation of min-heap operations
- Good use of helper methods (parent, leftChild, rightChild)
- Proper handling of heap properties and edge cases
- Well-commented code explaining the approach

Areas for Improvement:
- Add a destructor to delete the dynamically allocated heap array to prevent memory leaks
- Consider making the heap dynamically resizable instead of fixed capacity
- The printHeap method could be more informative by showing the complete heap structure
- Include error handling for cases like inserting into a full heap (currently just returns silently)
- The isLeaf method could be simplified to just check if i >= size/2 since i < size is always true in that case

PR #1280, opened 2025-08-19
Done competitivecoding1

ramya-gopalaswamy:


super30admin:
- Problem1.cpp: The solution is well-implemented and correct. No major improvements are needed.
- Problem2.cpp: The min-heap implementation is good, but consider using a dynamic array (like std::vector) instead of a fixed-size array to handle more elements. Also, adding error handling for edge cases (like extracting from an empty heap) would improve robustness. The code quality is high, but these changes would make it more flexible and production-ready.

PR #1290, opened 2025-08-26
Completed missing number in sorted array

meetrpatels30:


super30admin:
- The solution is correct and efficient, matching the reference solution in both time and space complexity.
- The code is clean and follows good practices. The use of binary search is appropriate for this problem.
- One minor improvement could be to add a comment explaining the logic behind the condition `arr[mid] == mid + 1` to make it even clearer for readers.
- The time complexity comment has a typo ("lon(n)" instead of "log(n)"), which should be corrected for clarity.

PR #1292, opened 2025-08-26
Missed element in sorted array

Saideepthi123:


super30admin:
Strengths:
- The code is well-structured and readable, with appropriate comments.
- The student correctly identified the need for a binary search approach and maintained O(1) space complexity.

Areas for Improvement:
- The logic used to determine the missing number is incorrect. The condition `if high-mid != nums[high] - nums[mid]` does not correctly identify the half where the missing number is located. Instead, you should compare the difference between the array value and its index (e.g., `nums[mid] - mid`) to determine where the missing number lies.
- The example usage and output comment are correct, but the function itself does not produce the correct result for the given input.

Suggested Correction:
To fix the logic, you should compare the difference between the array value and its index, similar to the reference solution. For example:
```python
def binary_missing(nums):
    low = 0
    high = len(nums) - 1

    while (high - low) > 1:
        mid = low + (high - low) // 2
        if (nums[low] - low) != (nums[mid] - mid):
            high = mid
        elif (nums[high] - high) != (nums[mid] - mid):
            low = mid

    return nums[low] + 1
```

PR #1294, opened 2025-08-28

vidyasreekumar:
Finding missing number in a sorted consecutive array using binary search

### DIFF
--- a/Problem1.java
+++ b/Problem1.java
@@ -1,1 +1,31 @@
+// Time Complexity : O(logn)
+// Space Complexity : O(1)
 
+// Perform binary search to check if a number is in the correct index
+// A number is in the correct index if the number = index + 1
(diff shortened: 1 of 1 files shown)


super30admin:
- The solution is well-implemented and correctly solves the problem. The binary search logic is correctly applied to find the missing number.
- The code is clean and follows good practices, such as using `low + (high - low) / 2` to avoid potential integer overflow.
- The comments are helpful but could be slightly more detailed to explain why the solution works (e.g., mentioning that the array is expected to start at 1 and be consecutive except for one missing number).
- The return statement `return arr[low]-1;` is correct but might benefit from a comment explaining why this works (i.e., because `low` ends up at the index where the missing number should be, and `arr[low]` is one more than the missing number).

PR #1296, opened 2025-09-04
Add MinHeap and missing number solution

ravithakur2k:
Introduced a MinHeap class with push, pop, top, and heapify methods in DesignMinHeap.py. Added a binary search-based solution for finding the missing number in a sequence in MissingNumber.py.

### DIFF
--- a/DesignMinHeap.py
+++ b/DesignMinHeap.py
@@ -1,0 +1,72 @@
+class MinHeap:
+    def __init__(self):
+        self.heap = [0]
+
+    def push(self, value):
+        self.heap.push(value)
+        i = len(self.heap) - 1
+
+        # Percolate up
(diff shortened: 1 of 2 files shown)


super30admin:
- Strengths:
  - Correct implementation of binary search for finding the missing number.
  - Good understanding of time and space complexity.
  - Clear and readable code structure.
- Areas for Improvement:
  - The `MinHeap` class, while correctly implemented, is not relevant to the problem and should be omitted.
  - Fix the typo in the `pop` method of `MinHeap` (`==` should be `=`).
  - The `MissingNumber.py` solution could be simplified by removing redundant checks (e.g., `elif mid + 1 != arr[mid]` is redundant since it's the opposite of the first condition).

PR #1301, opened 2025-10-06
Competitive-Coding-1 Complete

PrasiddhShah:


super30admin:
Strengths:
- Good understanding of binary search concept in Problem1.java.
- Proper handling of edge cases in Problem1.java.
- Code is well-commented and structured.

Areas for Improvement:
- In Problem1.java, the while loop condition should be `while (high - low > 1)` to avoid infinite loops. The updates to low/high should also be adjusted to ensure convergence.
- The minHeap implementation in Problem2.java has several issues:
  - The minHeapify method incorrectly swaps with the left or right child without comparing which is smaller.
  - The minHeap method has incorrect loop conditions and indexing.
  - The remove method does not handle the heap property correctly after removal.
- For Problem2.java, it would be helpful to study heap operations more thoroughly and implement them correctly.

PR #1310, opened 2025-10-25
Competitive-Coding-1

lalitha7571:
Competitive-Coding-1


super30admin:
- The solution is well-implemented and correctly solves the problem. The binary search approach is appropriate for this scenario.
- The variable naming and code structure are good, making the code easy to understand.
- One minor improvement could be to add a comment explaining why `high` is initialized to `n - 2` (since the array size is `n-1`), to make it even clearer for readers.
- The logic is concise and efficient, matching the reference solution in both time and space complexity.

PR #1316, opened 2025-11-21
"competative coding 1 done"

Sahithipsl470:


super30admin:
For the missing element problem:
- The logic for identifying the missing element is incorrect. Instead of averaging the values at low and high indices, you should return `ar[low] + 1` or `ar[high] - 1` based on the differences.
- The condition `if (lowDiff == highDiff)` is unnecessary and incorrect. The differences should not be equal if there is a missing element.

For the heap implementation:
- The heapify method should compare the current node with its children and swap if necessary to maintain the min-heap property. Your implementation incorrectly checks for larger values instead of smaller ones.
- The insert method has an infinite loop because the condition `while (size > 0)` is always true. It should be `while (curr > 0)`.
- The removeMin method should replace the root with the last element and then heapify from the root. Your implementation is mostly correct but relies on the flawed heapify method.

Overall, your code shows understanding of the concepts but needs improvement in implementation details and correctness.

PR #1319, opened 2025-12-29
Completed CP-1

rbhargav0104:


super30admin:
- Problem1.cpp:
  - Strengths: The binary search approach is a good choice for this problem.
  - Areas for Improvement: The logic for determining the missing number is incorrect. Consider comparing the difference between the element and its index, as in the reference solution. Also, add comments to explain the logic.

- Problem2.cpp:
  - Strengths: The structure of the minHeap class is clear, and the methods are appropriately named.
  - Areas for Improvement: The heapify-up and heapify-down logic is incorrect. Review the correct implementation of these operations. The `extractMin` method should use `h.back()` instead of `h.push_back()`. Add comments to explain the logic and ensure correctness.

PR #1328, opened 2026-02-02
1problem completed

BharathVuppala96:


super30admin:
Strengths:
- You have attempted to use binary search, which is the optimal approach for this problem.
- The code structure is simple and uses constant space.

Areas for Improvement:
- The variable 'n' is not defined. You need to know the expected maximum value (which is the length of the array + 1 since one number is missing). However, in the problem, the array has n-1 integers in the range 1 to n. So the length of the array is n-1, and n is len(arr)+1. You should compute n inside the method.
- The initial checks for the first and last element are good, but they should be implemented correctly. For example, if the first element is not 1, then 1 is missing. If the last element is not n (which is len(arr)+1), then n is missing. But you need to define n.
- The binary search logic needs to be revised. The idea is that in a complete sorted array from 1 to n, each element should be at index i-1 (if the array starts at index 0). So for any index i, the value should be i+1. If there is a missing number, then from the missing number onward, the value will be i+2 (if the missing number is before) or something else. Actually, the difference between the value and the index should be constant (1) until the missing number. After the missing number, the difference becomes 2. So you can use the difference to determine where the missing number is.
- The reference solution does: if (ar[a] - a) != (ar[mid] - mid), then the missing number is in the left half (a to mid). Otherwise, if (ar[b] - b) != (ar[mid] - mid), then it is in the right half (mid to b). Your condition `if arr[m]-m==1` is not equivalent. You should compare the difference at mid with the difference at the start or end.
- Also, your while condition is `while l<h`, which might not handle all cases. The reference solution uses `while (b - a) > 1`, which ensures that when the loop ends, a and b are consecutive indices. Then it returns ar[a] + 1.
- In your code, you return l+1. But if the array is [1,2,3,5], then l and h will converge to index 2 (value 3) and then you return 4? Actually, that might work for this case, but it's not general because your binary search logic is flawed.

Suggestions:
- Define n = len(arr) + 1 at the beginning.
- Check if the first element is not 1, return 1.
- Check if the last element is not n, return n.
- Then perform binary search by comparing the differences. You can use:
  low, high = 0, len(arr)-1
  while low <= high:
      mid = (low+high)//2
      if arr[mid] - mid == 1:
          low = mid+1
      else:
          high = mid-1
  Then return low+1? Actually, after the loop, low will be the index where the difference first becomes 2. So the missing number is low+1? But let's test with [1,2,3,5]: 
      Initially: low=0, high=3
      mid=1: arr[1]=2, 2-1=1 -> low=2
      mid= (2+3)//2 = 2: arr[2]=3, 3-2=1 -> low=3
      mid= (3+3)//2=3: arr[3]=5, 5-3=2 !=1 -> high=2
      Now low=3, high=2 -> exit. Then return low+1=4 which is correct.
  So this might work. But note: the condition should be if arr[mid] - mid == 1, then the missing number is to the right, else to the left.

  However, this is similar to your code but with a different while condition and updating high to mid-1. Your code updates h to m without subtracting one, which might cause an infinite loop.

- Alternatively, you can use the reference solution's approach which is more standard.

Revised code suggestion:

class Solution:
    def missingNumber(self, arr):
        n = len(arr) + 1
        if arr[0] != 1:
            return 1
        if arr[-1] != n:
            return n
        
        low, high = 0, len(arr)-1
        while low <= high:
            mid = (low + high) // 2
            if arr[mid] - mid == 1:
                low = mid + 1
            else:
                high = mid - 1
        return low + 1

But test with [1,2,3,5] as above: it works.
Test with [2,3,4,5] (missing 1): the initial check would catch that arr[0]!=1 -> return 1.
Test with [1,2,3,4] (missing 5): initial check arr[-1] != 5 -> return 5.

So this should work.

PR #1334, opened 2026-02-10
Mock Interview 1 | Missing element in a sorted array from a list of N…

sandy7907:
…atural numbers


super30admin:
Strengths:
- You have chosen an efficient algorithm (binary search) for the problem, which is appropriate for sorted arrays.
- The code is clean and uses standard binary search structure.

Areas for Improvement:
- The condition inside the binary search needs to be corrected. The key insight is that in a sorted array without duplicates and with one missing number, the difference between the value at an index and the index itself should be consistent until the missing number. For the first part of the array (before the missing number), the difference is 1 (since arr[i] = i+1). After the missing number, the difference becomes 2 (because the numbers are shifted by one). So, you should compare arr[mid] with mid+1. If arr[mid] == mid+1, then the missing number is to the right; otherwise, it is to the left.
- Consider using a condition like: if (arr[mid] != mid + 1), then the missing number is at or before mid. Otherwise, it is after mid.
- Also, ensure that the loop terminates correctly. The reference solution uses while ((b - a) > 1) to avoid infinite loops and to ensure that when the loop ends, the missing number is between a and b.
- Test your code with the provided examples. For the first example [1,2,3,5,6,7,8], the missing number is 4. Your code should return that.
- Add comments to explain the logic, especially the condition, to make the code more understandable.

Suggested corrected code:
```
int missingNumber(int[] arr) {
    int low = 0, high = arr.length - 1;
    while (low <= high) {
        int mid = low + (high - low) / 2;
        if (arr[mid] != mid + 1) {
            high = mid - 1;
        } else {
            low = mid + 1;
        }
    }
    return low + 1;
}
```
This corrected code will work because when the loop ends, low will be the index where the first inconsistency occurs, and the missing number is low+1.

super30admin:
Strengths:
- You have chosen an efficient algorithm (binary search) that is well-suited for this problem.
- The code is clean and follows a standard binary search structure.
- You used `low + (high - low) / 2` to avoid integer overflow, which is a good practice.

Areas for Improvement:
- The core logic of the condition inside the binary search is incorrect. The goal is to check whether the missing number is to the left or right of the current mid. The correct approach is to compare the value at the current index with the expected value. For an array starting at 1, the expected value at index `i` is `i+1`. So, if `arr[mid] == mid+1`, then the missing number is to the right; otherwise, it is to the left. Alternatively, you can use the difference: if `arr[mid] - mid` is greater than 1, then the missing number is on the left; otherwise, it is on the right. Your condition `arr[mid] - 2 == mid` does not generalize. For example, in the first test case [1,2,3,5,6,7,8], at index 3 (value 5), we have 5-2=3 which equals the index (3), so it would set high to mid-1, which might be correct in this case, but it fails for other cases. For instance, if the missing number is at the beginning, such as [2,3,4,5] (missing 1), your code would not work.

- Consider testing with edge cases: missing number at the beginning, end, or middle. For example:
  - Input [2,3,4,5] (missing 1): your code might not return 1.
  - Input [1,2,3,4] (missing 5): your code might not return 5.

- Suggested correction: Instead of `arr[mid] - 2 == mid`, use `arr[mid] != mid + 1` to determine if the missing number is at or before mid. However, note that if the missing number is at mid, then the value at mid is actually mid+2. So a common approach is to check the difference between the value and the index: if `arr[mid] - mid > 1`, then the missing number is on the left. Otherwise, it is on the right. You can adjust the binary search accordingly.

- Here is a corrected version for your reference:
  while (low <= high) {
      mid = low + (high - low) / 2;
      if (arr[mid] - mid > 1) {
          // missing number is to the left
          high = mid - 1;
      } else {
          low = mid + 1;
      }
  }
  return low + 1;  // because the array starts at 1, so the missing number is low+1? Actually, after the loop, low will be the index where the missing number should be. The expected value at index low is low+1, but if there is a missing number, the value at low is actually low+2. So returning low+1 is correct.

- However, test with the edge case when the missing number is the last one. For [1,2,3,4] (size 4, but n=5), the array has indices 0 to 3. The expected value at index i is i+1. The missing number is 5. The binary search would set low to 4 at the end, and return 5 (which is correct). Similarly, if the missing number is the first one [2,3,4,5] (n=5, array size 4), the expected value at index0 is 1, but we have 2. So the condition `arr[mid]-mid>1` at mid=0: 2-0=2>1 -> true, so we set high to -1, then low becomes 0, and we return low+1=1, which is correct.

- Therefore, the corrected condition is `if (arr[mid] - mid > 1)`, and then return `low+1`.

- Also, consider the case when the array has only one element. For example, if the array is [2] (n=2, missing 1), then the code should return 1. With the correction, it works: low=0, high=0, mid=0. arr[0]-0=2>1 -> true, so high becomes -1. Then return low+1=1.

- Another case: array [1] (n=2, missing 2). Then the condition: at mid=0, arr[0]-0=1>1? false, so low becomes 1. Then return low+1=2.

So the corrected code is:
```
int missingNumber(int[] arr) {
    int low = 0, high = arr.length - 1, mid;
    while (low <= high) {
        mid = low + (high - low) / 2;
        if (arr[mid] - mid > 1) {
            high = mid - 1;
        } else {
            low = mid + 1;
        }
    }
    return low + 1;
}
```

PR #1345, opened 2026-05-04
Solved find first missing number in an array

anirudhv98:
Completed competitive coding 1 - please review


super30admin:
### Interview Problem: Find Missing Number in a sorted array  (Problem1.cs)
Strengths:
- You have chosen an efficient algorithm (binary search) for this problem, which is appropriate given the sorted array and the O(log n) requirement.
- The code is clean and well-commented, making it easy to understand your thought process.
- You correctly calculated the mid-index without causing integer overflow by using `low + (high - low) / 2`.

Areas for Improvement:
- The main issue is with the condition `if (arr[mid] - mid > 1)`. This condition does not correctly identify where the missing number occurs. In a correct array without duplicates from 1 to n, each element should be equal to its index + 1. So if no number is missing, `arr[i] = i+1`. Therefore, `arr[i] - i` should be 1 for all i. When a number is missing, from the point of the missing number onward, `arr[i] - i` becomes 2 because the numbers are shifted by one. However, your condition checks if it is greater than 1, which might be true even if the missing number is not exactly at mid. Actually, the difference `arr[i] - i` should be compared to 1. If it is greater than 1, then the missing number is on the left. But note that it should be exactly 1 for all indices before the missing number and exactly 2 for indices from the missing number onward. So the condition should be: if `arr[mid] - mid > 1`, then the missing number is at or before mid. But actually, it's better to compare `arr[mid]` and `mid+1` (since the first element should be 1 at index0). Alternatively, you can use the reference solution's approach: compare the differences `ar[a] - a` and `ar[mid] - mid`. If they are not equal, the missing number is between a and mid.

- Another issue is that your code returns `result + 1`. This assumes that the missing number is always one more than the index where the condition is first met. But as seen in the example [2,3,4] (missing 1), your code returns 2 instead of 1. The correct missing number should be `arr[result-1] + 1` or something similar? Actually, the index where the missing number occurs is the first index where `arr[i] != i+1`. So if you find the first index i such that `arr[i] != i+1`, then the missing number is i+1. But in your code, you are setting result to mid when the condition is true, and then you return result+1. This might work if the missing number is not the first one, but fails for the first.

- Consider the reference solution: it uses two pointers and checks the differences. It stops when the gap between a and b is 1, then returns `ar[a] + 1`. This works because when the array has no missing number, `ar[i] - i` is constant. When there is a missing number, the array is split into two parts: left part where `ar[i] - i` is 1, and right part where it is 2. The reference solution finds the transition point.

- To fix your solution, you should adjust the binary search to find the first index i where `arr[i] != i+1`. Then the missing number is i+1. You can do this by checking if `arr[mid] == mid+1`. If yes, then the missing number is to the right (low = mid+1). Otherwise, it is to the left (high = mid-1) and you record that mid is a candidate. After the loop, the first index where it fails is low, so the missing number is low+1. But note: if the entire array is correct, then the missing number is n (the last one). So you need to handle that.

- Alternatively, you can use the reference solution's method which is more straightforward.

VERDICT: NEEDS_IMPROVEMENT

---

Your solution appears to be for a different problem (likely "Find the Missing Number" or similar) rather than the "Design Min Heap" problem. For the Min Heap problem, you need to create a class that represents a min heap with methods to insert, extract the minimum, and possibly get the minimum without removal. The solution should include:

- An array to store the heap elements.
- Methods to maintain the heap property (e.g., heapify up for insertion and heapify down for extraction).
- Proper handling of the heap structure (complete binary tree).

I recommend revisiting the problem statement and implementing the Min Heap from scratch. You can start by defining a class with a constructor that initializes the heap array. Then, implement methods for insert, extractMin, and getMin. Remember to use helper functions for parent, leftChild, rightChild, and swapping elements. Also, consider edge cases like an empty heap or full heap.

Your binary search solution is well-written for the problem it addresses, but it's crucial to read the problem carefully to ensure you're solving the correct one. If you have any questions about heap operations, feel free to ask for clarification.

VERDICT: NEEDS_IMPROVEMENT

